PR #6, opened 2024-09-27
Regenerate workflows and enable CI

mortenmj:
The workflow generation library in bb-storage needed some changes to get CI working. Also needed to update the lockfile and run gofumpt.

There are a lot of golint errors, which make this PR fail CI, that we’ll fix later


mortenmj:
Needs https://github.com/buildbarn/bb-storage/pull/222

PR #8, opened 2024-10-02
Let the frontend specify where the backend is via env vars

gempesaw:
## context

hey there! I'm interested in running bb-portal in other places besides `localhost` where it doesn't make sense for the frontend to be calling `localhost:8081`. Apparently (first time next dev), [nextJS has support for env vars](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser) via `.env` files, and particularly for getting vars all the way into client pages by using env vars prefixed with `NEXT_PUBLIC_`.

the resulting idea here is that I can run `NEXT_PUBLIC_BES_BACKEND_URL=http://localhost:9091 npm run dev` and the frontend will make its `/graphql` requests to that URL (or any arbitrary URL), so we're not locked into having the backend available on `localhost:8081`

## changes

- add two env vars for setting where the backend is.
  - the grpc URL one is just used in the helper text
  - defaults are the same as the current settings
- modify the frontend tsx to use those env vars

## testing

when running `npm run dev` without any changes, the UI looks the same, and the network requests are to :8081

image
image

and then i can set the env vars at run time and the frontend will pick up the values & hit `/graphql` elsewhere:

image
image


mortenmj (approved):
lgtm

PR #28, opened 2024-10-11
add _cgo back

trey-ivy:


mortenmj (approved):

patches/com_buildbarn_bb_storage/base_image.diff
```diff
@@ -29,4 +29,21 @@ index e4aa679..6e92429 100644
...
++        target_platform = "@rules_go//go/toolchain:linux_" + arch + "_cgo"
+         images.append(arch_image_target)
+         platform_transition_filegroup(
+             name = arch_image_target,
+@@ -49,6 +49,6 @@ def container_push_official(name, image, component):
+     oci_push(
+         name = name,
+         image = image,
+-        repository = "ghcr.io/buildbarn/" + component,
++        repository = "ghcr.io/buildbarn/" + component + "/backend",
```
mortenmj, line 47: Do we need this? I would have thought `ghcr.io/buildbarn/bb_portal` and `ghcr.io/buildbarn/bb_portal_frontend` would suffice

trey-ivy: seems neater to me.  maybe we can change frontend to match as well, ...but if you want to adjust in a subsequent PR I'm totally fine w/that.

PR #31, opened 2024-10-14
Add configuration proto and start gRPC/HTTP servers using config

mortenmj:
This uses the convenience methods from the rest of the Buildbarn ecosystem to launch the gRPC and HTTP servers. I'd rather not refactor to use proto config for everything, as part of one large PR, so for now we are in a transitionary state where we still have some of the old flags in use.


trey-ivy:
thanks for doing this morten!  I have some small nonblocking feedback.   Also I'm fine if you want to combine this and #32  into a single PR.  Seems cleaner to just cut over all the config to the jsonnet file at once, although at present I'm having issues building #32 locally.

trey-ivy (commented):
- add a sample jsonnet file somewhere
- update documentation on how to run
- handle use case where config file isn't supplied and provide feedback msg
- link to the open issue.

EdSchouten (commented):
LGTM!

PR #33, opened 2024-10-15
Remove blob download endpoint

mortenmj:
I don't think we need this functionality. It gives us the ability to download blobs from any gRPC CAS backend, but at the expense of unwanted complexity. If we want to provide the ability to download blobs from the UI at all (and we probably do), it seems reasonable to expect the user to also run bb-browser so that we can simply link to that.


mortenmj:
Had a conversation with @EdSchouten about the option of using bb-browser, and how that is not considered a public API we can depend on to be stable. After discussing with @trey-ivy we'll go ahead and remove this endpoint for now, and consider adding back similar functionality, but probably a different implementation, if the need arises.

PR #37, opened 2024-10-16
Inline cmd/bb_portal/portal-service.go

mortenmj:
Shouldn't need two files here. If we do, something should move to internal/ or pkg/


trey-ivy:
lgtm

PR #46, opened 2024-10-22
Add Source Control Tab for GitHub

trey-ivy:
Adds a tab for source control information that links to github and provides links to the following:

- Repository URL
- Branch Name
- Commit SHA
- Github Actor Profile
- Run ID for github actions
- Pull request

 Just did github for now, in the future we an look at other source control/ci systems as requested

Pasted Graphic 1


trey-ivy:
these environment variables should be available by default in the github actions CI context, so they shouldn't need to be set.

mortenmj (approved):

PR #50, opened 2024-10-24
Fix run id

trey-ivy:


gempesaw (approved):
heh, I was just looking at that runURL wondering where that magic number came from. good timing :)

(of course my approval has no power here 👍🏽 )

trey-ivy:
thanks @gempesaw yeah left over from testing and missed it :)

PR #51, opened 2024-10-25
Push Database Finish Processing Tasks to context background process

trey-ivy:
Currently, when you have a large build that takes severals seconds to save to the database, the process can fail to save if the client uses the `--bes_upload_mode=fully_async` [flag](https://bazel.build/reference/command-line-reference#flag--bes_upload_mode) 

This flag causes the client to send a graceful shutdown upon completion of transmission w/out waiting for an acknowledgment from the server.  If the database process isn't done saving before this shutdown is received, the context is destroyed and the save fails (see below)

image

This also causes issues when the server tries to acknowledge requests 

image

To address this, we can push the database save to a background process and let it finish in its own time, and we can ignore send errors when the client gracefully terminated the connection. (i think...? its not like we were doing anything but logging them anyway so....). 

- improve database save functionality to be more robust and store as much of an invocation as it can
- remove fk restraint on eventFile url, fix event file url formatting
- push db save to a background process
- remove caching for bazel invocations table view
- update make lint

### DIFF
--- a/internal/api/grpc/bes/server.go
+++ b/internal/api/grpc/bes/server.go
@@ -47,12 +47,32 @@ func (s BuildEventServer) PublishBuildToolEventStream(stream build.PublishBuildE
 	// We'll want to ack these once all events are received, as we don't support resumption.
 	seqNrs := make([]int64, 0)
 
-	ack := func(streamID *build.StreamId, sequenceNumber int64) {
+	ack := func(streamID *build.StreamId, sequenceNumber int64, isClosing bool) {
 		if err := stream.Send(&build.PublishBuildToolEventStreamResponse{
 			StreamId:       streamID,
 			SequenceNumber: sequenceNumber,
 		}); err != nil {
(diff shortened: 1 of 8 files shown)


internal/api/grpc/bes/server.go
```diff
@@ -47,12 +47,32 @@ func (s BuildEventServer) PublishBuildToolEventStream(stream build.PublishBuildE
...
 		}); err != nil {
-			slog.ErrorContext(stream.Context(), "Send failed", "err", err)
+
+			// with the option --bes_uplod_mode=fully_async or nowait_for_upload_complete
+			// its not an error when the send fails. the bes gracefully terminated the close
+			// i.e. sent an EOF. for long running builds that take a while to save to the db (> 1s)
+			// the context is processed in the background, so by the time we are acknowledging these
+			// requests, the client connection may have already timed out and these errors can be
+			// safely ignored
+			if isClosing && (err.Error() != "rpc error: code = Unavailable desc = transport is closing") {
```
mortenmj, line 62: Maybe define an error to compare to here?

mortenmj (approved):
lgtm

PR #56, opened 2024-11-13
Bytestream ephmeral links for downloading logs from the problems tab

trey-ivy:
At present, we surface an error when. bytestream url is encountered during blob serialization.  This PR adds a button to the output displayed on the problems tab for build and test errors.

While this solution is not perfect, it is at least a step towards a more complete implementation for handling bytestream data returned from a remote execution system for log output on errored steps.  Long term, we want a configuration option to inline the logs and allow people to chose whether or not to save items to the disk so that they will exist beyond the lifespan of the cached data, but for now, i've added a download button that will give you the ability to pull the blob from the cache for at least as long as its there, similar to how the profile download button is implemented

for build and test errors, a download button is provided when a bytestream url is found that points to the existing bb-browser public frontend url

the button has a tool tip and a warning icon and danger color so that people are notified ahead of time that this link is at best ephemeral and not to be depended on long term,

image

also added another status to the noop archiver as it was throwing an error on serialization, and put a docker command for launching sqllite browser in a shell script in the tools folder so i don't forget it and thought others may find it useful if they wish to explore the sqllite database that backs the application.


mortenmj (approved):
lgtm

PR #63, opened 2025-01-21
address performance issues w/passrate

trey-ivy:
when running in an environment where target entries scale and become very large, the pass rate functionality can cause the targets and tests views to become slow and unresponsive.  to address this issue, i'm removing pagination and pass rate from these views and adjusting the backend queries respectively to see if this improves performance enough to make the application more responsive.  I'm also adjusting the cache policy from cache and network to the default and adding a long poll interval.

### DIFF
--- a/config/portal.jsonnet
+++ b/config/portal.jsonnet
@@ -7,5 +7,6 @@
   grpcServers: [{
     listenAddresses: [':8082'],
     authenticationPolicy: { allow: {} },
+    maximumReceivedMessageSizeBytes: 10*1024*1024
   }],
 }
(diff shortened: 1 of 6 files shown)


trey-ivy:
new UI view doesn't look that much different, but its markedly more performant.  idk if it will be a big enough performance boost tho
image

trey-ivy:
the new views look different, but i don't think we're losing that much functionality.  idk how useful it was to have avg/min/max/passrate/#runs on the list view.  The idea was to implement sorting at some point to allow you to find things like your longest running jobs and most built targets/etc.  I'm still going to provide this data but it won't be pulled real time.  it will be calculated daily and moved to the trends view.  You can of course still write a custom query to pull the data from the graphql api, but the performance hit was causing views to break and become unresponsive.
 
image

PR #79, opened 2025-02-08
fix: actionCache miss reason does NOT have an "UNKNOWN" enum at index 0

jasonschroeder-sfdc:
The `MissReason` enum doesn't align with the missReason() label.

I also added "UNCONDITIONAL_EXECUTION" enum consistently (Green for the UI)

### DIFF
--- a/frontend/src/components/ActionCacheMetrics/ActionCacheMissTag.tsx
+++ b/frontend/src/components/ActionCacheMetrics/ActionCacheMissTag.tsx
@@ -13,14 +13,13 @@ import { Tag } from 'antd';
 import themeStyles from '@/theme/theme.module.css';
 
 export const ALL_STATUS_VALUES = [
-    'UNKNOWN',
     'DIFFERENT_ACTION_KEY',
     'DIFFERENT_DEPS',
     'DIFFERENT_ENVIRONMENT',
     'DIFFERENT_FILES',
     'CORRUPTED_CACHE_ENTRY',
     'NOT_CACHED',
-
+    'UNCONDITIONAL_EXECUTION',
 ] as const;
 export type MissDetailTuple = typeof ALL_STATUS_VALUES;
 export type MissDetailEnum = MissDetailTuple[number];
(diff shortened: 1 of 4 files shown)


frontend/src/components/ActionCacheMetrics/ActionCacheMissTag.tsx
```diff
@@ -13,14 +13,13 @@ import { Tag } from 'antd';
 import themeStyles from '@/theme/theme.module.css';
 
 export const ALL_STATUS_VALUES = [
-    'UNKNOWN',
```
trey-ivy, line 16: i think we should keep this value here.  This just provides something to display in case what we see here isn't something we were expecting.  It doesn't necessarily map back to anything from the BEP.

frontend/src/components/ActionCacheMetrics/ActionCacheMissTag.tsx
```diff
@@ -30,11 +29,6 @@ interface Props {
 }
 
 const STATUS_TAGS: { [key in MissDetailEnum]: React.ReactNode } = {
-    UNKNOWN: (
```
trey-ivy, line 33: again, i think we should leave this option to have something to default to in the event we can't match

pkg/summary/summary.go
```diff
@@ -48,7 +48,6 @@ func (r MissReason) EnumIndex() int32 {
 // String Enum helper method.
 func (r MissReason) String() string {
 	return [...]string{
-		"UNKNOWN",
```
trey-ivy, line 51: i think this is the big fix right here and an excellent catch

trey-ivy (changes requested):
Hey @jasonschroeder-sfdc 

Thanks for putting this up!  I have some minor feedback, but this is a great catch!

Also i think there is one more place you need to update.  frontend/src/components/ActionCacheMetrics/index.module.css should include the color for unconditional execution.

trey-ivy:
hey @jasonschroeder-sfdc 
 can you rebase on main?  there was an issue with the frontend docker file unrelated to your PR

frontend/src/components/ActionCacheMetrics/ActionCacheMissTag.tsx
jasonschroeder-sfdc: Reverted :)

frontend/src/components/ActionCacheMetrics/ActionCacheMissTag.tsx
jasonschroeder-sfdc: Reverted

jasonschroeder-sfdc:
All set, @trey-ivy ! I just picked green for `UNCONDITIONAL_EXECUTION` as it seems like we don't have any greens yet. But I have no UX talent and the color is arbitrary :)

jasonschroeder-sfdc:
Let me know if you'd like me to squash commits.

PR #85, opened 2025-02-12
Build Logs

trey-ivy:
This PR adds some functionality to capture build logs from the progress event, store them in the database, and then display them on a tab in the invocations view.

image

Bazel emits logs a screen at a time, so displaying them on a static page doesn't translate perfectly.  For instance if you `cat` the actual log file generated from a bazel build, the output is displayed (very quickly) one screen at a time, and then at the end you get a simple message w/just a few lines of detail.

https://github.com/user-attachments/assets/dbd33f80-552c-4f1c-a14a-5b96a18e0883

This naive first pass implementation just captures them, strips any empty/blank lines from the output and renders them with a react module that CSS formats ANSI encoded text.  Possible future improvements could allow for users to optionally configure the application to either store build logs or ignore them, streaming build logs in real time, and only pulling the logs from the database on tab load.  

Tested with some build logs that run into several MBs worth of data.  Seems to hold up okay.


trey-ivy:
local testing was fine but when i pushed this to our dev env i started getting some weird infinite loop behavior on the logs tab, probably due to react re-rendering the component, so i'm going to test this a bit more

PR #93, opened 2025-03-07
Add settings for controlling CORS and changing reverse proxy target

isakstenstrom:
I found when deploying bb-portal in some situations that the configurability of the application was lacking. The problems I had was that the it was nessecary to proxy the frontend through the backend due to CORS problems, and that the proxying was not configurable.

This PR:

- Moves the `allowedOrigins` configuration option from the gRPC-proxies to the root of the configuration, and also uses the setting to allow for CORS on the main HTTP server.
- Makes the target of the reverse proxy configurable

Note that this brings with it two breaking changes: 

1. The proxy has no default target, so if no target is configured, no proxy will be started. We could easily add back the old target as a default, but I didn't since I think this is a good way to disable the proxy. Let me know if you want to add a default.
2. I have configured the application to not allow CORS if `allowedOrigins` is not configured. I think this is better since we this way have secure defaults, but I can change the behavior to the reverse if you think that is better.


isakstenstrom:
@trey-ivy Friendly ping :)

trey-ivy (approved):

PR #97, opened 2025-03-21
Fix operations count not showing correctly

### LINKED ISSUE #94: Total number of operations count on scheduler page always 0
I'm connecting up bb-portal to a recent version of the scheduler and I'm always seeing the total operations count on the new scheduler page as zero even though the corresponding count on the scheduler admin page is non-zero.

Image

vs

Image

Is this is a configuration issue on my side? Incidentally all of the platforms always have a per-platform queue size of 0 in both UIs.

P.S. Love the new scheduler view, its nice to have a single view of the cluster.

isakstenstrom:
The design of `buildqueuestateproxy` is to only show content with instance names that the user is allowed to see, as specified by the `instanceNameAuthorizer`. The pagination info was intentionally set to 0 to now disclose information  about the part of the system that the user should not have access to. This caused to the total operations count to stop working.

This PR fixes the operations count by fetching all operations from the scheduler and recreating the pagination info based on the operations that the user is allowed to see.

Fixes #94

### DIFF
--- a/frontend/src/components/SchedulerGrid/index.tsx
+++ b/frontend/src/components/SchedulerGrid/index.tsx
@@ -1,6 +1,6 @@
 import { useGrpcClients } from "@/context/GrpcClientsContext";
 import { useQuery } from "@tanstack/react-query";
-import { Row, Space, Statistic, Typography } from "antd";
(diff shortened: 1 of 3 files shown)


trey-ivy (approved):
lgtm

PR #98, opened 2025-03-24
Allow empty instancenames

### LINKED ISSUE #95: Browser functionality does not work with an empty instance name prefix
Possibly related to #94

isakstenstrom:
Makes it possible to use bb-portal with an empty instance name. 

- Rewrite the component that generates links to the browser page to not reject an empty `instanceNamePrefix` and also include the `instanceNameSuffix` in the url.
- Allow for empty `instanceName`s in the url.  
- Rename `IsInstanceNamePrefixAllowed` -> `IsInstanceNameAllowed` to better reflect that it does not only operate on prefixes.
- Fix an outdated unrelated test

Closes #95


johnclark456:
This PR fixes the issue for me. Many thanks

isakstenstrom:
> question...does this routing schema work with repeated instance name prefixes? image

Yes, both in the scheduler view and the browser view.

PR #102, opened 2025-04-23
Prometheus/metrics

trey-ivy:
This PR creates a /metrics endpoint for prometheus to scrape.   It also adds a few basic dashboards to help users get started.  At present, metrics are only emitted for non-cached test executions and we also add a couple env vars that allow for configuration of how/when metrics are emitted.

Overview:
image

Targets
image

Tests
image

Eventually these dashboards will completely replace the Targets/Tests and Trends tabs in bbportal.  This type of data is better suited to a time series database vs. a relational db and the way the data is modeled by the ORM makes for performance issues as the targets tables grow quickly out of control.  Cascading deletes will fail to finish once the database reaches a certain size and it just unwieldy and slow because of these particular fields that contain so much data.  A single invocation can have 1/2 million targets or more.  This PR doesn't fully remove the functionality yet, but it changes the default to NOT store target and test data in bbportal.  You can still enable this behavior with the existing environment variables.

There are some nice things about how we display target and test data, and I would still like to surface target and test data with its richer metrics for a specific invocation.  The trouble comes from trying to add relationships to those objects. At present there is a blob that dumps the entire summary as a json object into a field on the invocation.  As a next step I plan to expose the target and test data stored there so that this functionality is not linked to the non-performant way we currently handle targets and tests in the bbportal application itself


mortenmj (approved):
lgtm

PR #103, opened 2025-05-09
Set uniform x-axis tick distance in the previous execution stats graph

isakstenstrom:
Previously, if you have platform queues with non-consecutive size classes, there would be gaps in the graph.


mortenmj (approved):
lgtm

PR #105, opened 2025-05-09
Get proto files for the frontend using bazel

isakstenstrom:
Previously, the proto files were manually added to the repo.


mortenmj (approved):
lgtm

PR #106, opened 2025-05-09
Consolidate endpoints

isakstenstrom:
Instead of having one endpoint for each gRPC-web proxy, the proxies are now integrated into the regular HTTP endpoint. It compares the beginning of the request paths to check if the request should sent to one of the proxies.

This also necessitated a minor refactor of the CORS handling.

This is a breaking change since the configuration for the proxies has been changed.


trey-ivy:
nice!  should simplify deployment.  lgtm

PR #108, opened 2025-05-14
Add `killOperationsAuthorizer`

isakstenstrom:
Adds a authorizer that controlls who can kill operations via bb-portal. Also hides the kill operations button in the frontend if the user isn't authorized to kill the operation, by querying a http endpoint on the backend.


mortenmj:
This also renames the grpcweb module to common. Should that perhaps be a separate PR, so that this one is a bit slimmer?

isakstenstrom:
Sure, I can split them into two RPs

isakstenstrom:
@mortenmj Now this PR is rebased on top of main and ready to go.

internal/api/common/common.go
```diff
@@ -3,11 +3,25 @@ package common
...
+func ExtractContextFromRequest(req *http.Request) context.Context {
+	var pairs []string
+	for key, values := range req.Header {
+		for _, value := range values {
+			pairs = append(pairs, key, value)
+		}
+	}
+	return metadata.NewIncomingContext(req.Context(), metadata.Pairs(pairs...))
+}
+
```
mortenmj, line 24: I'd like to see some tests for this

internal/api/grpcweb/buildqueuestateproxy/server.go
```diff
@@ -127,6 +131,39 @@ func (s *BuildQueueStateServerImpl) RemoveDrain(ctx context.Context, req *buildq
...
+
+	if s.IsAllowedToKillOperation(ctx, operationName) {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(`{"allowed": true}`))
+	} else {
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(`{"allowed": false}`))
+	}
+}
+
```
mortenmj, line 166: This should be tested

mortenmj (changes requested):
Overall I think this looks good, but it needs some tests

internal/api/grpcweb/buildqueuestateproxy/server.go
isakstenstrom: Done!

mortenmj:
Thanks for your contribution, @isakstenstrom!

PR #110, opened 2025-05-16
Use pagination on the backend to fetch ListOperations

isakstenstrom:
Currently, the Buildqueuestate proxy calls ListOperations with a page size of `math.MaxUint32` to fetch all operations (which it needs to do to apply the `InstanceNameAuthorizer` properly). This becomes a problem when the total number of operations is larger than the maximum response size that the grpc client can handle (4MB). 

This PR makes the proxy use pagination for its requests to avoid this. The page size is configurable with a recommended value of 500, which according to my crude experiments should be somewhere around 1MB.

I also added more tests for the function. I used mocking of the grpc client to test the proxy more properly. It uses the same mocking setup as bb-storage.


mortenmj (approved):
lgtm

PR #117, opened 2025-06-24
Use blobstore to fetch resources from storage, instead of grpc proxy

isakstenstrom:
Instead of just forwarding gRPC requests from the browser, bb_portal now handles the requests using blobstore. This is more configurable, has less code, and makes it possible to send requests directly to storage when using multiple storage instances, instead of having to go via a frontend instance.


meroton-benjamin (approved):
LGTM

Issue #23, opened 2024-10-11
Parity with other buildbarn services


trey-ivy:
we should have microservices match with the rest of the buildbarn project for consistency and completeness.  
- auth
- tracing
- jsonnet files for config
- etc

mortenmj:
Auth and tracing are now in place. We're using jsonnet/proto to configure some parts of the backend, not all configuration has been migrated just yet.

isakstenstrom:
@HagegeR Sorry, I updated how you supply the config file and forgot to update the documentation. Now that has been fixed.

Regarding the wiki: I agree and think we should move all documentation to the repo.

meroton-benjamin:
I'd say we're at parity now. If we can figure out something in particular we're missing feel free to open another ticket.

PR #128, opened 2025-07-21
Move configuration all to jsonnet and add more options

isakstenstrom:
This PR does the following things:

- Removes the `bep-folder` configuration option, as well as all code associated with it. From what we found, this code is never called so it can safely be removed (but we could be wrong).
- Moves all configuration options that were previously provided on the command line to the jsonnet file. This makes the jsonnet whole configuration for the backend. One breaking change is that there is no longer any defaults for any of the moved fields, but we have added the old defaults to the example configuration file.
- Changes the way the application expects the configuration file. The config file is no longer loaded with `--config-file <file>`. Instead, bb-portal expects the config file to be the first argument, and will return an error otherwise. This is the same behavior as the reset of Buildbarn.
- Adds options to disable manual BEP file upload and the GraphQL playground.

Fixes part of #23

### DIFF
--- a/pkg/proto/configuration/bb_portal/bb_portal.proto
+++ b/pkg/proto/configuration/bb_portal/bb_portal.proto
@@ -44,6 +44,9 @@ message BuildEventStreamService {
 
   // Enables the BEP file upload endpoint.
   bool enable_bep_file_upload = 4;
+
+  // Enables the Graphql playground.
+  bool enable_graphql_playground = 5;
 }
 
 message BrowserService {
--- a/cmd/bb_portal/main.go
+++ b/cmd/bb_portal/main.go
@@ -189,7 +189,9 @@ func newBuildEventStreamService(configuration *bb_portal.ApplicationConfiguratio
 	srv.Use(entgql.Transactioner{TxOpener: dbClient})
 
 	router.PathPrefix("/graphql").Handler(srv)
-	router.Handle("/graphiql", playground.Handler("GraphQL Playground", "/graphql"))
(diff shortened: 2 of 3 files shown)


pkg/proto/configuration/bb_portal/bb_portal.proto
```diff
@@ -10,9 +10,35 @@ import "pkg/proto/configuration/http/http.proto";
 
 option go_package = "github.com/buildbarn/bb-portal/pkg/proto/configuration/bb_portal";
 
+// Postgres database source type.
+message PostgresSource { string url = 1; }
```
meroton-benjamin, line 14: this is a connection string as understood by the pgx driver. We should link to the documentation https://pkg.go.dev/github.com/jackc/pgx/v5 and call this a a `connection_string`  to help with understanding what to enter here.

pkg/proto/configuration/bb_portal/bb_portal.proto
```diff
@@ -10,9 +10,35 @@ import "pkg/proto/configuration/http/http.proto";
...
+
+// SQLite database source type.
+message SqliteSource { string url = 1; }
```
meroton-benjamin, line 17: similarly this is a connection string as understood by the https://pkg.go.dev/github.com/mattn/go-sqlite3 package. They seem to call this both `connection_string` and `dsn` (data source name) in their docs.

meroton-benjamin (changes requested):

meroton-benjamin (approved):
LGTM

PR #169, opened 2025-11-25
Add `X-Requested-With` to the allowed headers otherwise the manual BEP upload fails with a CORS error

vtsao-openai:


isakstenstrom:
The change looks good, but I would like a commit message.

PR #205, opened 2026-03-10
dbcleanup: add a dedicated cleanup worker mode

sharmila-oai:
`bb-portal` currently starts the database cleanup loop inside every serving process. In a multi-replica deployment that multiplies cleanup work, makes compaction backlogs more expensive, and ties cleanup failures directly to the request-serving pods. We were seeing OOMs with especially trying to compact incomplete_build_logs.

Add a `--cleanup-worker-only` mode that starts only the BES cleanup loop. This lets deployments run cleanup in a separate singleton worker pod while leaving normal `bb-portal` replicas focused on serving traffic. I was thinking this could be a separate deployment and just re-use the backend image for bb-portal.

Also put a small ordered batch limit on `CompactLogs()`. A single cleanup pass should not try to compact every eligible invocation in one shot when there is a large backlog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,13 @@ bazel run //cmd/bb_portal -- $PWD/config/portal.jsonnet
 
 The backend runs a reverse proxy for the frontend.
 
+To run the cleanup loop in a dedicated worker process instead of the serving
+process, start the same binary in cleanup-worker mode:
+
+```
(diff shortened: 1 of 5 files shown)


isakstenstrom:
This is now possible via the regular config file, thanks to #355.

PR #227, opened 2026-04-21
Improve build page visualization

isakstenstrom:
NOTE: This is a stacked PR. Please merge #226 first.

This commit does the following:
* Shortens the invocation link in build details page, and adds the Bazel command as a column
* Creates a help function for converting a BazelInvocation object to a string
* Refactor the invocation timeline tooltip to no longer show the start/end time, and to also show the Bazel command, job, workflow, and action
* Improve x-axis tick rendering in the invocation timeline, which previously was placing the ticks on unexpected places
* Removes the link in the invocation timeline y-axis, and make the bar chart entries into links to the respective invocation

The Bazel command (both in the tooltip and table) is truncated if it's longer than 40 characters. The question is whether to render the canonical command instead, in which case we should update the commit before merging.

image
image


darkrift:
@isakstenstrom The images are not populating for me and returns a 404 when clicking on them.

isakstenstrom:
> @isakstenstrom The images are not populating for me and returns a 404 when clicking on them.

Whoopsie! It should be fixed now.

frontend/src/utils/commandLineDataToString.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const commandLineDataToString = (cmd: CommandLineData) => {
+  if (!cmd) return "unknown";
+  return [
+    cmd.executable,
+    cmd.command,
+    ...cmd.options.map(
+      (x: CommandLineOptions) => `--${x?.option}${x?.value && `=${x?.value}`}`,
+    ),
+    ...cmd.residual,
```
magnus2lin, line 14: Hi, thanks for the work on this PR really nice improvements overall

Testing PR236 in dev we couldn't open Builds from the main page, Firefox throws:
> Error: can't access property Symbol.iterator, e.residual is null

The cause is `cmd.residual` being null when invoking bazel query invocations where there's no residual section. I was able to get around that by adding a small guard `...(cmd.residual ?? [])`

To reproduce
```
echo '//...' > /tmp/targets.txt
bazel query ‑‑query_file=/tmp/targets.txt
```

isakstenstrom: > Hi, thanks for the work on this PR really nice improvements overall
> 
> Testing PR236 in dev we couldn't open Builds from the main page, Firefox throws:
> 
> > Error: can't access property Symbol.iterator, e.residual is null
> 
> The cause is `cmd.residual` being null when invoking bazel query invocations where there's no residual section. I was able to get around that by adding a small guard `...(cmd.residual ?? [])`
> 
> To reproduce
> 
> ```
> echo '//...' > /tmp/targets.txt
> bazel query ‑‑query_file=/tmp/targets.txt
> ```

Good catch!

I elected to fix this in #226 instead by making sure that the backend sends and empty array for the `Options`, `StartupOptions`, and `Residual` instead of sending nil. This aligns the response with the generated typescript types that don't mark them as optional. Please give it a try!

magnus2lin: Super nice, thanks @isakstenstrom! I tested this in our dev environment and can confirm it resolves the issue for us.

meroton-benjamin (approved):

PR #239, opened 2026-04-29
feat: surface per-test logs in Tests tab (PLAT-1102)

aron-muon:
## Summary

- Ingest `TestResult.test_action_output` from BEP events into the existing `invocation_files` table, linked to `TestResult` via a new optional ent edge
- Expose the per-test files through GraphQL as `TestResult.actionOutputs`
- Add an expandable row to the invocation Tests tab so users can click through to `test.log`, `test.xml`, and per-attempt logs without leaving bb-portal

## Why

Today bb-portal's Tests tab only shows status/cached/duration. The actual test output (`test.log`, `test_attempts/attempt_*.log`) is uploaded to bb-storage CAS by the remote executor, and the BEP `TestResult` event contains `bytestream://` URIs to those blobs — but `save_test_result.go` reads every other field on the event and silently drops `GetTestActionOutput()`. Verified empirically against staging: `invocation_files` for a recent failing-test invocation contained only `command.profile.gz`, `coverage_report.lcov`, and three summary metric entries. No test logs.

So users hit a failed test and have no way to see the output without re-running locally or adding `--test_output=errors` to `.bazelrc`.

### Schema (commit 1)

Adds an optional bidirectional edge between `InvocationFiles` and `TestResult`. Existing invocation-scoped files (BuildToolLogs, profile, coverage) continue to set only the `bazel_invocation` edge — the new column is nullable and indexed.

### Backend ingestion + GraphQL (commit 2)

- New ` saveTestActionOutputs` runs after `createTestResultsBulk` in the same transaction. It re-queries the just-inserted `TestResult` rows by their natural key `(label, configId, run, shard, attempt)`, then bulk-inserts `InvocationFiles` rows with `OnConflict().DoNothing()` so replayed BES events don't abort the batch on the `(name, bazel_invocation)` unique constraint.
- Files without a digest are skipped (they couldn't be fetched from CAS anyway).
- New `TestActionOutput` GraphQL type and `TestResult.actionOutputs` field. Resolver walks the new edge and skips files missing digest metadata. Mirrors the existing `Profile` resolver pattern.
- Reuses existing `besfileutils.go` helpers (`getFileDigestFromBesFile`, etc.) so URI parsing stays consistent with `saveBuildToolLogs.go`.

### Frontend (commit 3)

- Extends `GET_TESTS_FOR_INVOCATION` to fetch `testResults { id, run, shard, attempt, status, actionOutputs { name, digest, sizeInBytes, digestFunction } }`.
- New `ExpandedRow` component groups outputs per test result (sorted by run/shard/attempt) and renders each output as a link to the existing `/api/v1/servefile/{instanceName}/blobs/{digestFunction}/file/{hash}-{sizeBytes}/{name}` endpoint.
- `rowExpandable` is gated on `actionOutputs` presence — cached test hits or invocations whose CAS objects have been evicted simply have no expand chevron rather than an empty panel.

## Test plan

- [x] `go generate ./...` succeeds; `sql/migrations/schema.sql` diff is just one new nullable bigint column + index on `invocation_files`
- [x] `go vet ./internal/database/buildeventrecorder/ ./internal/graphql/` clean
- [x] `cd frontend && npm run build` passes (16/16 routes, no type errors)
- [ ] Deploy to staging
- [ ] Trigger a fresh CI run with a known failing test (e.g. `//tests/aggregation:test_aggregator`)
- [ ] Open the invocation in staging UI → Tests tab → click expand on the failing row → confirm `test.log` and `test_attempts/attempt_*.log` appear with sizes
- [ ] Click a link, confirm the log content renders in the browser
- [ ] Confirm passing/cached tests have no expand chevron (rowExpandable returns false)
- [ ] Sanity-check staging DB:
      ```sql
      SELECT i.name, i.test_result_test_action_outputs IS NOT NULL AS linked
      FROM invocation_files i
      JOIN bazel_invocations bi ON bi.id = i.bazel_invocation_invocation_files
      WHERE bi.invocation_id = '<new-invocation-id>'
        AND i.name LIKE '%/testlogs/%';
      ```
      should return non-empty results with `linked = true`


aron-muon:
Apologies — opened against the wrong repo. This is for our internal fork. Re-opening against aron-muon/bb-portal.

PR #240, opened 2026-05-08
frontend: wrap long text in pre elements

AlexTereshenkov:
I have noticed that in a couple of text blocks, long lines were not wrapped leading to text getting out of the panels. Please see the images attached (`Failure Message` panels):

image

image

There are a few more places where no style is applied, but I am not seeing any poorly rendered text blocks. If I discover this happens in other places, I'll submit a separate PR.

I believe this style change should be robust, but please advise if there's better way to achieve prettier formatting in `pre` elements.


AlexTereshenkov:
@isakstenstrom Could you please approve the workflows run? Many thanks!

isakstenstrom:
Looks good! Thanks for the contribution!

PR #251, opened 2026-05-19
Mark go_deps overrides as dev_dependency

ma-oli:
Split the gazelle go_deps extension proxy into two: a dev-only go_deps_dev that holds gazelle_override and module_override calls (which are forbidden in non-root modules), and a regular go_deps that keeps from_file and use_repo so consumers still see the repo aliases.


meroton-benjamin (changes requested):
The change in MODULE.bazel looks good to me but where do the changes to MODULE.bazel.lock come from? Are those accidental pull ins from some local changes and if not we should look at adding a CI step to reflect the requirement.

meroton-benjamin:
solved in #274

PR #14, opened 2024-10-10
Cleanup database relationships, add trends view and test summary/details views

trey-ivy:
This PR starts adding options to explore build metrics across all invocations.  

## Trends
It adds a trends view, which displays data for all invocation durations

image

and plans are to add a few additional visualizations to this view.

And it also adds a couple new views for tests.

## Tests Summary

A searchable grid of all your unique test targets.
image
paginated for larger data sets.
expanding rows provides a link to the most recent runs.  you can click the buttons and navigate straight to each invocation
image

## Test Details
Provides more data and a further lookback period, and also a graph of duration over time so you can see if tests suddenly start taking longer to run to help spot regressions.
image
again, plans to expand on this view with further visualizations in the future.

## Database Relationships
Also as mentioned in #11 there are several database relationships that need attention.  This PR addresses everything I could find there.  Additionally it extends the custom resolvers to add some queries used for exploring test data and stubs out a few methods for doing something similar for target data in the future.  

## Branding
I've also added a few options for branding.  if you optionally include a company name and/or slack channel environment variables they'll be displayed  in the header, footer and title page.

image

image

image

## Other Minor updates

- sort invocations table by most recent first
- link tests to their given overviews on the invocation page/ tests tab
- nil checking for cancelled invocations

PR #252, opened 2026-05-20
Fix empty Trends chart: XAxis dataKey points to nonexistent field

walter-zeromatter:
## Problem

On the Trends page, the invocation-durations chart renders only the YAxis (often with a "first tick" in the hundreds of thousands of milliseconds) and no Area. The summary stats (Total / Avg / Median / Max / Min) render correctly.

## Root cause

`frontend/src/components/pages/Trends/index.tsx` declares:

```tsx
<XAxis dataKey="name" />
```

But `dataPoints` only contains `invocationId`, `from`, `to`, and `duration` — there is no `name` field. So the XAxis dataKey resolves to `undefined` for every row.

This has been the code since the original Trends PR (#14), but it rendered correctly under recharts v2: a category XAxis with an undefined dataKey silently fell back to row-index positioning. After commit 256ecfa ("Build frontend with Vite instead of Nextjs") bumped recharts from `^2.12.7` to `^3.8.0`, the v3 axis logic collapses all points to a single undefined category bucket, so the Area degenerates and the chart appears empty. The YAxis is unaffected and still auto-scales to the duration range — hence the lonely "450000"-style tick.

## Fix

Point XAxis at `from` (the invocation start timestamp) so points are chronologically ordered, and format both axes for readability:

- X ticks as locale dates
- Y ticks via `readableDurationFromMilliseconds` (smallestUnit: `"s"`)

## Verification

- `npx biome check` clean
- `npx tsc --noEmit` clean
- Have not yet loaded the rebuilt frontend in a browser — happy to attach a screenshot if a maintainer can point me at the preferred local-dev path, or I can verify against our internal deployment and report back.

### DIFF
--- a/frontend/src/components/pages/Trends/index.tsx
+++ b/frontend/src/components/pages/Trends/index.tsx
@@ -166,8 +166,17 @@ export const TrendsPage: React.FC = () => {
                       <stop offset="95%" stopColor="#8884d8" stopOpacity={0} />
                     </linearGradient>
                   </defs>
-                  <XAxis dataKey="name" />
-                  <YAxis />
+                  <XAxis
+                    dataKey="from"
+                    tickFormatter={(v: string) =>
+                      new Date(v).toLocaleDateString()
+                    }
+                  />
+                  <YAxis
+                    tickFormatter={(v: number) =>
+                      readableDurationFromMilliseconds(v, { smallestUnit: "s" })
+                    }
+                  />
                   <CartesianGrid vertical={false} strokeDasharray="3 3" />
                   <Tooltip />
                   <Area


meroton-benjamin:
Is the trends view something you actually use? It is mostly a placeholder and I think we should instead remove it completely until we can add something useful there.

walter-zeromatter:
It's something I'd like to use, yes. I was hoping to add some additional data there as well, but even just seeing a trend of invocation durations would be helpful.

Once that's working I was going to try and add a few others as needed to help me tune my deployment. Cache hit rate, build success/fail rate, worker utilization, response times, etc.

Some of that I can get from Grafana + metrics but It'd be nice to be able to drill in & see the invocation ids that had low cache hit rates for example right from the same site

meroton-benjamin (approved):

meroton-benjamin:
Sure, I guess we can hold off on getting rid of it as it can easily be disabled with configuration, but as it is right now it's not a very meaningful view.

PR #269, opened 2026-05-28
Fixes uint64 to int erroneous conversion

darkrift:
This was created using codex.

It creates a new UnsignedLong type that can be used end to end from the database to graphql using numeric as the database representation.

Added a new manual test to load a BEP trace along with the frontend to manually navigate to it and inspect the result using `bazel run //test/manual/unsigned_long`

Fixes #194


isakstenstrom:
Are you sure we need to do all that? I think we should be fine just annotating the Ent fields with this:
```
field.Uint64("bytes_sent").Optional().
	SchemaType(map[string]string{
		dialect.Postgres: "NUMERIC(20, 0)",
	}),
```

darkrift:
@isakstenstrom the problem is that the Int graphql type is limited to a signed 32 bits : https://spec.graphql.org/draft/#sec-Int

```
The Int scalar type represents a signed 32-bit numeric non-fractional value. Response formats that support a 32-bit integer or a number type should use that type to represent this scalar.
```

This original change was making the uint64 type to be supported end to end, not just the database mapping

darkrift:
We should probably map them as `UnsignedLong` though : https://scalars.graphql.org/chillicream/unsigned-long.html

darkrift:
@isakstenstrom any thoughts on this ? I can rework this out because of all the conflicts if we agree on that.

isakstenstrom:
I'm not convinced that this complexity is worth it for solving this very small problem. The problem will only occur when Bazel sends something larger than Int64 max, which shouldn't happen if Bazel is doing things correctly. So it kinda is a non-problem.

Would using `UnsignedLong` make this a lot less complex? In that case it could maybe be okay.

darkrift:
> Would using UnsignedLong make this a lot less complex

Using UnsignedLong where ? 

There is no such type for go or postgres.

The graphql is defining that as an UnsignedLong already which is a custom type taking a string as input and output it as a bigint in javascript because js doesn't also have unsignedlong, only number and bigint.

isakstenstrom:
> We should probably map them as `UnsignedLong` though : https://scalars.graphql.org/chillicream/unsigned-long.html

Oh I probably misunderstood you. I assumed this was talking about another way to implement this using Graphql UnsignedLong, but that is the implementation you are already using, right?

PR #271, opened 2026-06-03
Default `created_timestamp` to insertion time

### LINKED ISSUE #270: Auto-migrate fails adding bazel_invocations.created_timestamp to non-empty databases
Deploying bb-portal against a Postgres database from an existing bb-portal deployment that predates commit d527dbf ("Add CreatedTimestamp to invocation") fails at startup with:
```
Fatal error: rpc error: code = Unknown desc = Failed to create BES service:
Could not automatically migrate to desired schema:
sql/schema: modify "bazel_invocations" table:
ERROR: column "created_timestamp" of relation "bazel_invocations"
contains null values (SQLSTATE 23502)
```

###  Root cause

bazel_invocations.created_timestamp was added as `timestamptz NOT NULL` with no `DEFAULT` clause. When Ent's auto-migrate (`Schema.Create` in `cmd/bb_portal/main.go`) tries to `ALTER TABLE bazel_invocations ADD
  COLUMN created_timestamp timestamptz NOT NULL` against a table with existing rows, Postgres rejects the statement.

The field is set by application code at insert time (internal/database/buildeventrecorder/build_event_recorder.go uses time.Now()), but that doesn't help legacy rows that already exist when the new binary starts up.

### Reproduction
1. Deploy any bb-portal commit from before d527dbf against a Postgres DB and add at least one Bazel invocation (a row in bazel_invocations).
2. Upgrade to any commit at or after d527dbf (2026-04-20).
3. The new binary fails to start with the error above.
  
### Workaround (operator-side)

My quick fix to get this working in my deployment was to  manually ALTER TABLE to add the column nullable, backfill the `created_timestamp` to the value of the existing `started_at` column, then set NOT NULL — before deploying:
```sql
BEGIN;
ALTER TABLE bazel_invocations ADD COLUMN created_timestamp timestamptz;
UPDATE bazel_invocations
  SET created_timestamp = COALESCE(started_at, now())
  WHERE created_timestamp IS NULL;
ALTER TABLE bazel_invocations ALTER COLUMN created_timestamp SET NOT NULL;
COMMIT;
```
Using COALESCE(started_at, now()) preserves real-event ordering for legacy rows; started_at is the closest semantic match to "time the invocation was first inserted." Rows that never produced a BuildStarted event fall back to now().

### Proposed fix

It seems to be like a good approach could be to add both a default to the schema field:

```go
  field.Time("created_timestamp").
      Default(time.Now).
      Immutable().
      Annotations(
          entgql.Skip(),
          entsql.Default("CURRENT_TIMESTAMP"),
      ),
```

With this approach, legacy rows backfilled by Postgres would get stamped with the migration time, not their original insertion time. 
Alternatively, it could use `entsql.DefaultExpr` to set the `created_at` timestamp to the value of another row, like `started_at`. I'm not sure which is the best option here.

Beyond that, is there a best approach I can take as an operator to manage migrations for schema changes like this?

matx-sdaniel:
Resolves #270
The `created_timestamp` field on `BazelInvocation` was added as a NOT NULL column with no default. When deploying against a database that predates the field, Ent's auto-migrate fails with
```
ERROR: column "created_timestamp" of relation "bazel_invocations"
contains null values (SQLSTATE 23502)
```
because Postgres rejects `ALTER TABLE ... ADD COLUMN ... NOT NULL` with no default against a non-empty table.

Add `.Default(time.Now)` and `entsql.Default("CURRENT_TIMESTAMP")` so the column descriptor carries both an application-side default for the Ent insert builder and a DB-level default that backfills existing rows at migration time. Drop the now-redundant `SetCreatedTimestamp` call from the test helper.

The big decision is if the current timestamp at the time of the migration is acceptable, or if it's better to use DefaultExpr or DefaultFunc to backfill a more meaningful timestamp.


meroton-benjamin:
Hi,
Thank you for your contribution but at its current version bb-portal doesn't perform database migrations besides the initial bootstrapping one and any migrations which ent happens to be capable of resolving automatically. As the data is temporary log data consumers are encouraged to scrap the database in between schema incompatible versions. When we do add migrations we want to do that at bootstrapping time in a more structured manner so that a client can properly recognize that its understanding of the database is outdated, that the migrations can be verified properly in CI and deal with merging older migrations properly.

PR #273, opened 2026-06-19
testresult: persist TestResult.test_action_output, surface as download links in TestDetails

matx-lai:
## Summary

Today the bb-portal frontend shows test status, timing, and a hostname for each test result, but not the actual artifacts the test produced. The references to those artifacts (`test.log`, `test.xml`, `test.outputs/outputs.zip`, etc.) are already in the BES `TestResult.test_action_output` field, they just aren't persisted or surfaced.

This PR persists them and renders a small "Outputs" column on the test detail page with one link per artifact. URIs are passed through the existing `/blob?uri=...` proxy so `bytestream://` URIs become downloadable.

This unblocks a common debugging workflow: clicking through a failed test to grab its undeclared outputs (`outputs.zip`) directly from the UI. In deployments where bb-portal is the only CAS-facing UI (no separate `bb-browser`), this is currently impossible from the browser — you have to shell out and fetch the blob by digest.

## What changed

- `ent/schema/testresult.go` — new `test_action_outputs` JSON field, stored as `map[string]any` keyed by output name. Same encoding shape as the existing `timing_breakdown` field so it surfaces through the generated `Map` GraphQL scalar without needing a new `entgql.Type`.
- `internal/database/buildeventrecorder/save_test_result.go` — extracts `TestResult.test_action_output` and flattens it into the new column (`{name, uri, digest, length}` per entry). Marshals to `null` for empty so consumers don't see `{}`.
- `sql/queries/test_result.sql` + `sql/migrations/schema.sql` — bulk-insert plumbing for the new column.
- `frontend/src/components/pages/TestDetails/` — new `TestActionOutputs.tsx` component, wired into `columns.tsx` as a new "Outputs" column. Renders "—" for empty.
- Generated files (`ent/gen`, `sqlc`, `gqlgen`, frontend GraphQL codegen) refreshed.

## Codegen

Ran the equivalent of `tools/update_schema.sh`:

```
go generate
bazel run //:gazelle
bazel run //tools:sqlc -- generate
(frontend) npm run generate-graphql
```

`bazel run //cmd/bb_export_schema` hit a pre-existing `ape+/assimilate:tar must produce a single file` transitive dep error on my checkout (reproduces on a clean baseline without these changes), so the `sql/migrations/schema.sql` diff was applied by hand. The added column is `"test_action_outputs" jsonb NULL` — the same form `bb_export_schema` emits for other JSON columns. Happy to regenerate it if a maintainer can reproduce the `bb_export_schema` build.

## Test plan

- [x] `bazel build //cmd/bb_portal` succeeds.
- [x] `npm run check-types` and `npm run build` in `frontend/` succeed.
- [ ] Manual: point a bb-portal at a BES stream containing test events, confirm the new column renders links and clicking through downloads the expected blob.
- [ ] Reviewer to extend `save_test_result_test.go` if more coverage is wanted on the BES → JSON shape.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


matx-lai:
Could a maintainer kindly approve workflow runs on the lint-fix commit (5f6ed80)? The previous commit's lint failure was a `gofmt` skew in two generated resolver files — collapsing four consecutive `string` params on `GetTarget`, and folding ~56 individual `type fooResolver struct{ *Resolver }` decls into a single `type ( … )` block. `gofmt -l` is clean on both files locally, so the lint job should go green once the workflows actually run.

cc @mortenmj — would you have time to take a look when CI is green?

isakstenstrom:
I'm already working on a change that will store and display output files for targets, tests and actions, as a continuation of #241. We want to display the output files for all 3, and for it to have good performance and save space, we need a common good way to store those files. I don't see anyting wrong with this implementation, but I suggest that we wait for my implemtnation, which should be done soon™.

matx-lai:
@isakstenstrom Is there an ETA on when https://github.com/buildbarn/bb-portal/pull/241 (or equivalent) might be merged?

isakstenstrom:
> @isakstenstrom Is there an ETA on when #241 (or equivalent) might be merged?

Sorry, I've been busy with other things. The implementation is mostly done, but we are not sure that it will hold up when the volume of invocations increase (since we are handling all "NamedSetOfFiles" messages as well). We are going to see if we can optimize it somehow and do some performance testing to see if it works. I will try to publish a draft PR next week, but it could be quite a while before it can be merged.

isakstenstrom:
@matx-lai Now my PR is up(#300). Please take a look at it to see if it satisfies you needs.

We ended up splitting my work into two parts. One part that saves the test action output (#300) and one part that saves NamedSetOfFiles (like in #241) that is still a work in progress.

isakstenstrom:
#300 is merged. @matx-lai Please check if the merged version works for you.

PR #274, opened 2026-06-19
bb_portal: make bb_portal importable as a module dependency

ma-oli:
Importing bb_portal as a module dependency from another project currently results in an error as go_deps cannot be overridden in non-root projects. We're fixing that here by making go_deps a dev_dependency, as it is typically done in most projects.


meroton-benjamin (approved):
Thanks for the contribution

PR #278, opened 2026-06-25
frontend: fix vite build chdir when bb_portal is an external module

ma-oli:
The vite_build rule passed `JS_BINARY__CHDIR = package_name()` ("frontend") to the vite launcher. That path only exists when bb_portal is the main repo; when bb_portal is consumed as an external Bazel module the sources live under `bazel-out/.../bin/external/<repo>/frontend`, so the launcher fails with "cd: frontend: No such file or directory".

Derive the chdir (and the `../` depth used for --outDir) from the output directory, which carries the `external/<repo>` prefix in that case. This is behavior-preserving for bb_portal's own build (pkg_dir = bin/frontend -> chdir "frontend", num_up 4) and lets the embedded frontend build when bb_portal is imported as a dependency.


isakstenstrom:
Hey!

#276 should have solved the same problem. Please check if it solves your usecase.

PR #308, opened 2026-08-02
Add a hermetic local bb-portal stack using rules_itest

darkrift:
### Summary

This PR provides a complete local bb-portal development and remote-execution environment managed by `rules_itest` and Docker Compose.

The stack includes:

- Independent Vite frontend and bb-portal backend services.
- PostgreSQL.
- A complete Buildbarn RBE path: storage frontend, two storage shards, scheduler, worker, and runner.
- Jaeger for traces from bb-portal and every Buildbarn component.
- Persistent PostgreSQL, CAS, action-cache, tracing and worker volumes.
- Health checks covering PostgreSQL, CAS, worker registration, and Jaeger.

Docker must be running. Container images are pinned by digest for reproducibility.

### Development workflow

For fast feedback, use [iBazel (bazel-watcher)](https://github.com/bazelbuild/bazel-watcher). iBazel watches a Bazel command's inputs and reruns affected targets when files change, bringing the hot-reload workflow familiar from JavaScript development servers to any Bazel target.

Start the stack with:

```sh
ibazel run --config=enable_reload //test/itest:bb_portal
```

Combined with `rules_itest` per-service reload:

- Vite handles browser hot-module reload.
- Backend changes rebuild and restart only bb-portal.
- PostgreSQL and Buildbarn remain running unless their Compose inputs change.

A one-shot launch without file watching is also available:

```sh
bazel run --config=enable_reload //test/itest:bb_portal
```

The `enable_reload` configuration disables frontend embedding in the backend. bb-portal instead proxies the standalone Vite service.

### Local remote execution

The `local_rbe` configuration connects Bazel to the local Buildbarn instance and selects a Linux/amd64 execution platform compatible with the worker.

It also selects Linux-compatible host/bootstrap tools, preventing macOS binaries such as `protoc` or Go builders from being sent to Linux workers.

Run the included smoke action and publish its BEP to bb-portal:

```sh
bazel build \
  --config=local_rbe \
  --config=bb_portal_itest \
  --noremote_accept_cached \
  //test/itest:rbe_smoke
```

The same configurations can be used with regular repository targets.

### Tracing

bb-portal and all Buildbarn components export OTLP traces to Jaeger:

- bb-portal
- storage frontend
- storage shards
- scheduler
- worker
- runner

The action runner is attached to a dedicated internal tracing network. It can reach Jaeger's OTLP receiver without gaining access to the host or external network.

Bazel itself does not currently create the root OpenTelemetry trace; its invocation ID can be used for future correlation with BES data.

### Endpoints

| Service | Address |
|---|---|
| Vite | http://127.0.0.1:5173 |
| Buildbarn scheduler admin | http://127.0.0.1:7982 |
| Buildbarn RE/CAS/AC | `grpc://127.0.0.1:8980` |
| Buildbarn build-queue state | `grpc://127.0.0.1:8984` |
| Jaeger | http://127.0.0.1:16686 |
| OTLP/gRPC | `grpc://127.0.0.1:4317` |
| OTLP/HTTP | http://127.0.0.1:4318 |
| PostgreSQL | `127.0.0.1:15432` |
| bb-portal UI/API | http://127.0.0.1:18081 |
| bb-portal BES | `grpc://127.0.0.1:18082` |

### Data lifecycle

Stopping iBazel does not delete the persistent database or Buildbarn caches.

Remove the containers while retaining their volumes:

```sh
bazel run //test/itest:buildbarn_compose -- down
```

Remove the containers and all local PostgreSQL, CAS, and action-cache data:

```sh
bazel run //test/itest:buildbarn_compose -- down --volumes
```

### Additional changes

- Use `rules_docker_compose` 0.3.0 from the Bazel Central Registry.
- Pin `rules_itest` to commit `107f75ea3d73bad8c61cebaa68c98c8c92e87672` for per-service reload support.
- Replace the dedicated embedded-PostgreSQL itest helpers with Compose-managed PostgreSQL.
- Disable generated hygiene tests because the development stack intentionally uses fixed ports.
- Add a stack health-check tool that waits for PostgreSQL, CAS shards, worker registration, and Jaeger.
- Add a tool-free remote-execution smoke action.
- Avoid forwarding a stale `Content-Length` header when proxying the Vite frontend, with regression coverage.
- Retain the embedded-database cleanup improvements used by other tests.
- Document startup, RBE usage, endpoints, tracing, and volume cleanup.

### Verification

- `bazel run //tools:reformat`
- `bazel test --config=enable_reload //pkg/frontend:frontend_test`
- `bazel build --config=enable_reload //test/itest:bb_portal`
- Verified PostgreSQL, CAS, worker registration, and Jaeger through the stack health check.
- Verified `//cmd/bb_portal` compiles and links through the local Linux worker, with 2,192 remotely executed actions.
- Verified Buildbarn component traces are visible through the Jaeger API.


darkrift:
@isakstenstrom Could you review this ?

isakstenstrom:
Am I correct in thinking that the database will be wiped when restarting the backend? I don't think that is the best for local development, as then you have to repopulate the database every time you need to test something.

Wouldn't a docker compose file be easier to work with? That is what we are using internally. I could maybe look at upstreaming our dev setup.

darkrift:
Yes but that's also a nice property when doing lots of database changes.

You can achieve the db repopulation easily with saving your BEP to an ndjson and import it afterward in the main page of bb-portal. I've been using that method extensively to repopulate it when adding new fields/tables.

If you want to connect to an external db instead of spinning up a short live postgres instance, we can add a flag and select on this to provide the pg connection string as an env var to use in the portal config instead of being hardcoded and not spin up a postgres instance at all if the provided value is different than the default connection string

darkrift:
FYI, rules_itest has this PR to provide native support for external services : https://github.com/hermeticbuild/rules_itest/pull/69

meroton-benjamin:
So this is an attempt to start a fully self contained local development stack, this is in and of itself a good thing to have. However the issue I see with this is that it does not capture enough of the development stack. It starts a postgres server using our embedded postgres setup (which was intended for use in tests but kudos on finding another use of it) and starts bb-portal with the frontend resources embedded into the binary itself.

In our dev setup (which we have not upstreamed) the way we do things is that we setup the local development stack by having a bootstrap script with docker-compose. This starts all the supporting services needed (postgres, jaeger, bb-storage, bb-scheduler etc) and we then run bb-portal itself with bazel run configured in such a manner that it forwards its frontend calls to a local development vite server.

That is not a fully hermetic stack, the portal then depends on these non-bazel managed resources existing simultaneously. The upside is that the local development vite server can do things such as hot reloading and instant builds of frontend resources and we also have the ability to actually run remote executions builds and see how the integration behaves.

So while we could add this development setup it would be an incomplete development setup and the project would then have two setups to maintain, with two different config files, two different port configurations etc etc.

How do you feel about us trying unify this instead and documenting the preferred dev setup for other contributors?

darkrift:
https://github.com/bazelbuild/bazel-watcher can do about the same with any bazel target to hot reload on code update, bringing the vite hot-reload feature to any bazel target :)

Also, rules_itest supports it (https://github.com/hermeticbuild/rules_itest#note and it now needs an equal sign for flags with value such as "--config=reload-config" instead of "--config reload-config"). That will keep the postgres server running and not "cleaned" if other components are hot reloaded

We could configure it the same way you have for the bb-portal to be 2 different process, the frontend and the backend separated so that when either is modified, they are rebuilt individually. Even improving your experience, because now the backend could also be hot reloaded with ibazel

Then, we could enhance this setup to include more components to it, or probably combine that with rules_docker_compose ?

That said, the bep ndjson is enough so far to be able to work with without a whole buildbarn stack, and these could be added in another flavor of the dev setup to have a full stack or a minimal stack.

darkrift:
Example of ibazel (bazel-watcher) output when it detect changes :
```

goroutine 1 [running]:
main.must(...)
	external/rules_itest+/cmd/svcinit/main.go:35
main.main()
INFO: Analyzed target //test/itest:bb_portal (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //test/itest:bb_portal up-to-date:
  bazel-bin/test/itest/bb_portal
INFO: Elapsed time: 18.986s, Critical Path: 17.04s
INFO: 6 processes: 6 action cache hit, 2 internal, 3 darwin-sandbox, 1 local.
INFO: Build completed successfully, 6 total actions
INFO: Runnable command line: bazel-bin/test/itest/bb_portal test/itest/portal.jsonnet
iBazel [9:26AM]: IBAZEL BUILD SUCCESS
iBazel [9:26AM]: Error writing success to stdin: write |1: broken pipe
iBazel [9:26AM]: using default bazel_fix_commands
INFO: Analyzed target //test/itest:bb_portal (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //test/itest:bb_portal up-to-date:
  bazel-bin/test/itest/bb_portal
INFO: Elapsed time: 18.986s, Critical Path: 17.04s
INFO: 6 processes: 6 action cache hit, 2 internal, 3 darwin-sandbox, 1 local.
INFO: Build completed successfully, 6 total actions
INFO: Runnable command line: bazel-bin/test/itest/bb_portal test/itest/portal.jsonnet

iBazel [9:26AM]: using 0 matching commandlines
iBazel [9:26AM]: Changed: "/Users/rlavoie/rlavoie/darkrift/bb-portal/pkg/invocation/actionanalytics/analyze.go". Rebuilding...
iBazel [9:26AM]: Running //test/itest:bb_portal
iBazel [9:26AM]: Notifying of changes
iBazel [9:26AM]: Error writing build to stdin: write |1: broken pipe
INFO: Analyzed target //test/itest:bb_portal (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //test/itest:bb_portal up-to-date:
  bazel-bin/test/itest/bb_portal
INFO: Elapsed time: 1.684s, Critical Path: 0.15s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Runnable command line: bazel-bin/test/itest/bb_portal test/itest/portal.jsonnet
iBazel [9:26AM]: IBAZEL BUILD SUCCESS
iBazel [9:26AM]: Error writing success to stdin: write |1: broken pipe
iBazel [9:26AM]: using default bazel_fix_commands
INFO: Analyzed target //test/itest:bb_portal (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //test/itest:bb_portal up-to-date:
  bazel-bin/test/itest/bb_portal
INFO: Elapsed time: 1.684s, Critical Path: 0.15s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Runnable command line: bazel-bin/test/itest/bb_portal test/itest/portal.jsonnet
```

darkrift:
Thinking about it even more, with rules_itest and having all that bazel managed for the bb setup, we could even have more complete integration tests that would also include your upcoming change for CompletedActionsLogger

darkrift:
@isakstenstrom @meroton-benjamin Can you check if that setup better suits you now ?

It sets frontend and backend as indivudual components and brings a buildbarn rbe stack along with it under docker compose including postgres and jaeger.

The volumes are persistent, but one can still delete them to start anew

internal/database/embedded/embedded_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestCleanupRemovesTemporaryDirectories(t *testing.T) {
+	runtimePath := t.TempDir()
+	dataPath := t.TempDir()
+	databaseProvider := &DatabaseProvider{
+		runtimePath: runtimePath,
+		dataPath:    dataPath,
+	}
+
+	if err := databaseProvider.Cleanup(); err != nil {
```
meroton-benjamin, line 17: require.NoError(t, databaseProvider.Cleanup())

internal/database/embedded/embedded_test.go
```diff
@@ -0,0 +1,25 @@
...
+		t.Fatalf("Cleanup failed: %v", err)
+	}
+	for _, path := range []string{runtimePath, dataPath} {
+		if _, err := os.Stat(path); !errors.Is(err, os.ErrNotExist) {
```
meroton-benjamin, line 21: require.Equal

pkg/frontend/frontend.go
```diff
@@ -105,6 +106,8 @@ func setupProxyHandler(router *mux.Router, sourceConfig *bb_portal.FrontendServi
 			}
 
 			r.Body = io.NopCloser(bytes.NewReader(newIndexContent))
+			r.ContentLength = int64(len(newIndexContent))
```
meroton-benjamin, line 109: Why was this added?

test/itest/buildbarn/config/storage.jsonnet
```diff
@@ -0,0 +1,59 @@
...
+
+{
+  grpcServers: [{
+    listenAddresses: [':8981'],
+    authenticationPolicy: { allow: {} },
+  }],
+  maximumMessageSizeBytes: common.maximumMessageSizeBytes,
+  global: common.global('bb-storage'),
+  contentAddressableStorage: {
+    // Prebuilt remote tools such as Node exceed the per-block limit of a
```
meroton-benjamin, line 42: What are you referring to with this block layout? Simply that we need a larger disk (than 4GB?) to fit the node binary? If the block size is an issue consider setting the new/spare/current/old to 1/1/3/1 to increase the block size instead of increasing the storage.

test/itest/buildbarn/config/worker-hardlinking-ubuntu22-04.jsonnet
meroton-benjamin, line 1: bb-portal also visualizes the fuse worker setup

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
+services:
+  volume-initializer:
+    image: docker.io/library/postgres@sha256:d3e1620b530c944afa6e887d22eb899824da68e19c52024bf98f5220c88a65b2
```
meroton-benjamin, line 3: * the volume initializer should not be a postgres container (even though I recognize that this is an "optimization" since we are already using the same postgres below)
 * avoid using docker.io its aggressive rate limits makes it likely not to work in many environments where a public ip is shared
 * use semantic tags rather than sha256 tags for this (i.e. a tag pointing to a major version etc)

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+      - storage-ac-1:/volumes/storage-ac-1
+      - storage-cas-0:/volumes/storage-cas-0
+      - storage-cas-1:/volumes/storage-cas-1
+      - storage-fsac-0:/volumes/storage-fsac-0
+      - storage-fsac-1:/volumes/storage-fsac-1
+      - jaeger-data:/volumes/jaeger
+      - worker-hardlinking:/volumes/worker
+
+  postgres:
+    image: docker.io/library/postgres@sha256:d3e1620b530c944afa6e887d22eb899824da68e19c52024bf98f5220c88a65b2
```
meroton-benjamin, line 38: docker.io and semantic tags

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+      retries: 30
+      start_period: 5s
+      timeout: 5s
+    ports:
+      - 15432:5432
+    volumes:
+      - postgres-data:/var/lib/postgresql/data
+
+  jaeger:
+    image: docker.io/jaegertracing/jaeger@sha256:46a886260e04002d8f45e213fc39063fa11a50446048fdaa64786fc0840cb9f8
```
meroton-benjamin, line 56: docker.io and semantic tags

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+      - tracing
+    volumes:
+      - ./config/jaeger.yaml:/etc/jaeger/config.yaml:ro
+      - jaeger-data:/badger
+    depends_on:
+      volume-initializer:
+        condition: service_completed_successfully
+
+  frontend:
+    image: ghcr.io/buildbarn/bb-storage@sha256:36201141f924865ddbea1bc518c4f98dd0dbb744181ef34f9f5e864d3dcf43f8
```
meroton-benjamin, line 73: semantic tags

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+    command: [/config/frontend.jsonnet]
+    ports:
+      - 8980:8980
+    volumes: &config-volumes
+      - ./config/common.libsonnet:/config/common.libsonnet:ro
```
meroton-benjamin, line 78: You are pulling in all the config files here while you should really only be pulling in two

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+      - ./config/worker-hardlinking-ubuntu22-04.jsonnet:/config/worker-hardlinking-ubuntu22-04.jsonnet:ro
+    depends_on:
+      jaeger:
+        condition: service_started
+      scheduler:
+        condition: service_started
+      storage-0:
+        condition: service_started
+      storage-1:
+        condition: service_started
```
meroton-benjamin, line 92: these dependencies are not necessary, the services are loosely coupled

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+
+  storage-0:
+    image: ghcr.io/buildbarn/bb-storage@sha256:36201141f924865ddbea1bc518c4f98dd0dbb744181ef34f9f5e864d3dcf43f8
```
meroton-benjamin, line 95: semantic tag (which for bb components is already hash suffixed)

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+    command: [/config/storage.jsonnet]
+    volumes:
+      - ./config/common.libsonnet:/config/common.libsonnet:ro
+      - ./config/storage.jsonnet:/config/storage.jsonnet:ro
+      - storage-ac-0:/storage-ac
+      - storage-cas-0:/storage-cas
+      - storage-fsac-0:/storage-fsac
+    depends_on:
+      jaeger:
```
meroton-benjamin, line 104: remove jaeger coupling

test/itest/buildbarn/docker-compose.yml
```diff
@@ -0,0 +1,200 @@
...
+    image: ghcr.io/buildbarn/bb-storage@sha256:36201141f924865ddbea1bc518c4f98dd0dbb744181ef34f9f5e864d3dcf43f8
+    command: [/config/storage.jsonnet]
+    volumes:
+      - ./config/common.libsonnet:/config/common.libsonnet:ro
+      - ./config/storage.jsonnet:/config/storage.jsonnet:ro
+      - storage-ac-1:/storage-ac
+      - storage-cas-1:/storage-cas
+      - storage-fsac-1:/storage-fsac
+    depends_on:
+      jaeger:
```
meroton-benjamin, line 119: remove jaeger coupling

test/itest/README.md
```diff
@@ -0,0 +1,72 @@
...
+PostgreSQL, Jaeger, and a complete local Buildbarn remote-execution path in
+Docker Compose. Docker must be running. The first start pulls
+the pinned Buildbarn and Ubuntu runner images, so it takes longer than later
+starts.
+
+For development, start the stack with iBazel so only the service whose inputs
+changed is restarted:
+
+```sh
+~/.local/bin/ibazel run --config=enable_reload //test/itest:bb_portal
```
meroton-benjamin, line 13: assuming of course that there is where you have ibazel installed, maybe link to where to get ibazel and assume user will put it on path?

test/itest/BUILD.bazel
```diff
@@ -0,0 +1,118 @@
...
+itest_service(
+    name = "bb_portal_backend",
+    args = ["$(location :portal.jsonnet)"],
+    data = [":portal.jsonnet"],
+    exe = "//cmd/bb_portal",
+    expected_start_duration = "5s",
+    http_health_check_address = "http://127.0.0.1:18081/",
+    hygienic = False,
+    shutdown_timeout = "15s",
+    tags = ["requires-network"],
```
meroton-benjamin, line 89: I don't think this build target requires the network

test/itest/BUILD.bazel
```diff
@@ -0,0 +1,118 @@
...
+
+itest_service(
+    name = "bb_portal_frontend",
+    exe = "//frontend:vite_dev_server",
+    # The first Vite dependency optimization pass takes longer than restarts.
+    expected_start_duration = "15s",
+    http_health_check_address = "http://127.0.0.1:5173/",
+    hygienic = False,
+    shutdown_timeout = "15s",
+    tags = ["requires-network"],
```
meroton-benjamin, line 77: I don't think this build target requires the network

test/itest/BUILD.bazel
```diff
@@ -0,0 +1,118 @@
...
+    health_check_args = [
+        "postgresql://app:password@127.0.0.1:15432/postgres?sslmode=disable",
+        "127.0.0.1:8980",
+        "127.0.0.1:8984",
+        "http://127.0.0.1:16686/api/services",
+    ],
+    health_check_timeout = "10m",
+    hygienic = False,
+    shutdown_timeout = "45s",
+    tags = [
```
meroton-benjamin, line 63: I'm not familiar with itest_services, but l don't think this build target requires the network, I don't think requires-docker is a recognized keyword by bazel, and if it was I don't this build target requires it.

requires-network is an instruction to bazel that when setting up the local action sandbox to allow that sandbox to reach the network, it doesn't propagate when doing e.g. bazel run or using this target as a dependency.

test/itest/BUILD.bazel
```diff
@@ -0,0 +1,118 @@
...
+        ":buildbarn_service",
+    ],
+)
+
+itest_service_group(
+    name = "bb_portal",
+    hygienic = False,
+    services = [":bb_portal_backend"],
+    tags = [
+        "requires-docker",
```
meroton-benjamin, line 101: I don't think this build target requires the network

test/itest/portal.jsonnet
```diff
@@ -0,0 +1,102 @@
+// Standalone configuration for //test/itest:bb_portal.
+//
+// The fixed loopback ports intentionally make it easy for a separate Bazel
+// invocation to stream its BEP to this service:
+//   PostgreSQL: 127.0.0.1:15432
```
meroton-benjamin, line 5: why are we adding 1 before the port numbers here?

test/itest/buildbarn/config/worker-hardlinking-ubuntu22-04.jsonnet
```diff
@@ -0,0 +1,42 @@
...
+      endpoint: { address: 'unix:///worker/runner' },
+      concurrency: 4,
+      instanceNamePrefix: 'hardlinking',
+      platform: {
+        properties: [
+          { name: 'OSFamily', value: 'linux' },
+          { name: 'container-image', value: 'docker://ghcr.io/catthehacker/ubuntu:act-22.04@sha256:dd7654ffb01d5b7b54b23b9ce928a1f7f2d08c7b3d7e320b6574b55d7ccde78b' },
+        ],
+      },
+      workerId: {
```
meroton-benjamin, line 27: you don't need to match the worker id format from bb-deployments you just need a unique id

test/itest/buildbarn/config/common.libsonnet
```diff
@@ -0,0 +1,81 @@
+{
+  blobstore: {
+    contentAddressableStorage: {
+      sharding: {
```
meroton-benjamin, line 4: It doesn't really seem necessary to have sharding for a dev environment, we can probably simplify this quite a bit

test/itest/buildbarn/config/common.libsonnet
```diff
@@ -0,0 +1,81 @@
...
+          stringValue: serviceName,
+        },
+      }],
+      sampler: {
+        always: {},
+      },
+    },
+    diagnosticsHttpServer: {
+      httpServers: [{
+        listenAddresses: [':80'],
```
meroton-benjamin, line 73: We probably don't want to use privileged ports (<1024) even if we in this context is using docker for this.

test/itest/buildbarn/config/storage.jsonnet
```diff
@@ -0,0 +1,59 @@
...
+    oldBlocks: 8,
+    currentBlocks: 24,
+    // AC and FSAC objects may be updated in place, which requires a single
+    // writable generation. This setting is also valid for the CAS.
+    newBlocks: 1,
+    blocksOnBlockDevice: {
+      source: {
+        file: {
+          path: path + '/blocks',
+          sizeBytes: blocksSizeBytes,
```
meroton-benjamin, line 22: this is not the size of a block it's the size of the file, it can just be "storageSizeBytes"

meroton-benjamin (changes requested):
I haven't had time to test the ergonomics of using this but it looks like it should cover most existing use cases. I left some comments on the code itself and will try to get time to feel this out on our end.

pkg/frontend/frontend.go
darkrift: This is a legit bug, newIndexContent is being appended injectFrontendConfigScript() which injects new content, but the content length was never updated.

test/itest/BUILD.bazel
darkrift: That's exactly the use case I'm using this for, so that I can run tests with a properly setup bb-portal + the buildbarn stack. This will require the network.

test/itest/buildbarn/docker-compose.yml
darkrift: This should be fixed, however I kept the sha256 tags when used for tests but added the label to make it more identifiable.

rules_itest allows you to run full integration tests which is what I temporarily used in other PRs. That would be useful to add long term to properly test bb-portal end to end

test/itest/buildbarn/docker-compose.yml
darkrift: They are not so loosely coupled, the frontend needs them when that setup is used to run tests and making sure the services are up and running before the bb-portal can accept data.

test/itest/portal.jsonnet
darkrift: This is to prevent clashing with the current setup of a user, if that user already has a postgres running on his host

test/itest/BUILD.bazel
darkrift: I moved the tag on the test target directly

darkrift:
@meroton-benjamin I did a pass on the comments and added a couple fixes

PR #309, opened 2026-08-02
Persist and display extended Bazel metrics with safe concurrent imports

darkrift:
## Summary

  Expand the invocation Metrics page to persist and display additional metrics emitted in Bazel’s Build Event Protocol. 

  ## Changes

  - Persist and expose additional metrics through Ent, PostgreSQL, and GraphQL:
      - Package counts and per-package load metrics
      - Bazel server cumulative analysis/build counts
      - Skyframe evaluation statistics
      - Rule-class and aspect build-graph counts
      - Dynamic execution race results
      - Persistent worker details, memory samples, and worker IDs
      - Worker-pool lifecycle statistics
      - Action-cache semaphore wait time
      - Critical-path duration

  - Add Metrics page panels and sortable tables for:
      - Package and Bazel server metrics
      - Package load metrics
      - Build-graph evaluation, rule-class, and aspect metrics
      - Dynamic execution races
      - Persistent workers and worker pools

  - Place “Package and Bazel Server Metrics” beside “Garbage Collection Breakdown” in a responsive row.
  - Support Bazel versions using either the legacy singular worker ID or newer repeated worker IDs.
  - Handle optional metrics safely. Sections omitted by older Bazel versions remain absent, while missing scalar values use protobuf/UI zero defaults.

  - Add an extended BEP fixture and GraphQL golden coverage for the new metrics.
  - Ignore local .ijwb workspace metadata.

  ## Database impact

  This adds new metrics tables, relationships, and columns to the Ent/PostgreSQL schema. Generated Ent models, SQL schema, GraphQL schema, resolvers, and frontend GraphQL types are updated accordingly.

  ## Testing

  - bazel test //internal/database/buildeventrecorder:buildeventrecorder_test
  - bazel test //test/integrationtest:integrationtest_test
  - bazel build //frontend:dist
  - npm run check-types

## Screenshots 
Screenshot 2026-08-02 at 5 36 12 PM

Screenshot 2026-08-02 at 5 36 33 PM

Screenshot 2026-08-02 at 5 36 57 PM


ent/schema/workerpoolmetrics.go
isakstenstrom: Since this entity has no fields of its own, it can be removed and the foreign of the metrics entity can be added directly to the `WorkerPoolStats`

isakstenstrom:
Hey @darkrift!

Me and @meroton-benjamin has discussed this and worked out another approach for how to handle the metrics for a invocation. Instead of creating a table for each type of metric, we could just store the metrics as a json blob in the database and let the frontend parse it. That would save us from having to do all this work of setting up database entities for all possible types of metrics and it would allow us to do this type of changes without changing the database.

As it is right now we don't use the metrics in any cross-invocation application, so there is no need to have it as separate tables. It would also fix #269 automatically.

What do you think about such a change?

darkrift:
@isakstenstrom doing so, you would block the opportunity for bb-portal (or other tools) to compute cross-invocation or cross target statistics and trends using the database with their proper field types. 

While you don't have an application for this right now, this is something that would really be nice to have eventually.

I have another PR in the works to display the actions done as part of an invocation with sums and avg, stddev, etc that needs stats to be computed server side because actions are paged and the stats would not be accurate if computed by the frontend.

darkrift:
@isakstenstrom there seems to be an increasing need for those metrics to be properly modelled so they can be extracted for metrics purposes through prometheus : https://bazelbuild.slack.com/archives/CA31HN1T3/p1787146586251829?thread_ts=1787145580.257629&cid=CA31HN1T3

isakstenstrom:
> @isakstenstrom doing so, you would block the opportunity for bb-portal (or other tools) to compute cross-invocation or cross target statistics and trends using the database with their proper field types.
> 
> While you don't have an application for this right now, this is something that would really be nice to have eventually.
> 
> I have another PR in the works to display the actions done as part of an invocation with sums and avg, stddev, etc that needs stats to be computed server side because actions are paged and the stats would not be accurate if computed by the frontend.

Alright, we don't have any such plans, but if you do we should keep the current structure.

Could you please split the target creation stuff into another PR, it has very little to do with the metrics?

internal/database/buildeventrecorder/saveBuildMetrics.go
```diff
@@ -298,25 +413,264 @@ func (r *buildEventRecorder) saveTargetMetrics(ctx context.Context, tx *ent.Clie
...
+	}
+	if len(nonNilPackageLoadMetrics) == 0 {
+		return nil
+	}
+
+	err := tx.PackageLoadMetrics.MapCreateBulk(nonNilPackageLoadMetrics, func(create *ent.PackageLoadMetricsCreate, i int) {
+		packageLoadMetric := nonNilPackageLoadMetrics[i]
+		loadDurationInNs := int64(0)
+		if packageLoadMetric.LoadDuration != nil {
+			loadDurationInNs = packageLoadMetric.LoadDuration.AsDuration().Nanoseconds()
```
isakstenstrom, line 451: Isn't it better to let the database field be null than to default to 0 for this field?

internal/database/buildeventrecorder/saveBuildMetrics.go
```diff
@@ -298,25 +413,264 @@ func (r *buildEventRecorder) saveTargetMetrics(ctx context.Context, tx *ent.Clie
...
+	return nil
+}
+
+func (r *buildEventRecorder) saveWorkerMetrics(ctx context.Context, tx *ent.Client, workerMetrics []*bes.BuildMetrics_WorkerMetrics, metricsDbID int64) error {
+	for _, workerMetric := range workerMetrics {
+		if workerMetric == nil {
+			continue
+		}
+
+		create := tx.WorkerMetrics.Create().
```
isakstenstrom, line 600: It would be nice if this could use batch inserting, to reduce the number of database requests. Or do we know that the number of worker metrics will be small?

internal/database/buildeventrecorder/saveBuildMetrics.go
```diff
@@ -193,10 +195,123 @@ func (r *buildEventRecorder) saveBuildGraphMetrics(ctx context.Context, tx *ent.
...
+	}
+	err = r.saveBuildGraphAspectCounts(ctx, tx, buildGraphMetrics.Aspect, buildGraphMetricsDb.ID)
+	if err != nil {
+		return util.StatusWrap(err, "Failed to save build graph aspect counts")
+	}
+	return nil
+}
+
+func (r *buildEventRecorder) saveBuildGraphRuleClassCounts(ctx context.Context, tx *ent.Client, ruleClassCounts []*bes.BuildMetrics_BuildGraphMetrics_RuleClassCount, buildGraphMetricsDbID int64) error {
+	nonNilRuleClassCounts := make([]*bes.BuildMetrics_BuildGraphMetrics_RuleClassCount, 0, len(ruleClassCounts))
```
isakstenstrom, line 219: I like the way you filter out `nil` before bulk inserts.

In my opinion, a nicer way to do this would be to do it the same way in for example `recordInvocationTags` in `internal/database/buildeventrecorder/save_structured_command_line.go` where we create a builder array that we populate in a loop. But this is a bit nitpicky, I'm fine with you current solution.

frontend/src/components/DynamicExecutionMetricsDisplay/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+interface Props {
+  dynamicExecutionMetrics: BazelInvocationMetricsDynamicExecutionMetricsFragment;
+}
+
+type RaceStatistic = NonNullable<
+  BazelInvocationMetricsDynamicExecutionMetricsFragment["raceStatistics"]
+>[number];
+
+const columns: TableColumnsType<RaceStatistic> = [
+  { title: "Mnemonic", dataIndex: "mnemonic" },
```
isakstenstrom: Don't use the `dataIndex` field when specifying columns in Ent tables. It sidesteps type checking and makes it hard to ensure that correct field names are used. Please use `render: (_, record) => record.mnemonic,` and add a key for each column instead..  This applies for all tables.

frontend/src/components/BuildGraphEvaluationMetricsDisplay/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+      evaluated: 0,
+    };
+    const countField = OPERATION_FIELDS[stat.operation ?? ""];
+    if (countField) {
+      row[countField] += stat.count ?? 0;
+    }
+    rowsBySkyfunction.set(skyfunctionName, row);
+  }
+
+  const rows = [...rowsBySkyfunction.values()].sort(
```
isakstenstrom: `rows` is not that descriptive when we have 3 tables in the same component.

internal/database/buildeventrecorder/saveBuildMetrics.go
darkrift: I prefer to have non-null values for metrics in the database because of the potential computing of value without requiring explicit null checks. Plus there is really no difference between 0 and null in this case is there ?

internal/database/buildeventrecorder/saveBuildMetrics.go
darkrift:
```
buildMetrics.workerMetrics represents persistent worker processes on the local Bazel host, not remote-execution hosts.

  Examples include local Javac, Kotlin, or Scala persistent workers launched by Bazel’s worker strategy. Its fields include the local process ID, memory usage, worker IDs, sandbox mode, and actions executed (third_party/bazel/
  protobuf/build_event_stream.proto:1224).
```

I don't think this is going to be quite high, I rebuilt a whole java project without remote exec (but still using remote cache) using `--strategy=Javac=worker` and I got these stats from our invocations:
```
 ### Current invocation

  - Result: success
  - Duration: 12m 7.331s
  - Total actions: 17,095
  - Worker actions: 1,260 (7.371%)
  - Worker metric records: 1
  - Worker IDs: 5
  - Worker-stat samples: 1
  - Worker-pool records: 1
  ```

Unless having persistent workers implemented remotely is a thing ?

frontend/src/components/BuildGraphEvaluationMetricsDisplay/index.tsx
darkrift: done

frontend/src/components/DynamicExecutionMetricsDisplay/index.tsx
darkrift: done

isakstenstrom (approved):
Looks good! Just needs a rebase to fix the merge conflicts and I can merge it.

darkrift:
@isakstenstrom done !

PR #312, opened 2026-08-05
Stamp image tags from STABLE_ status keys

matx-sjacob:
Related https://github.com/buildbarn/bb-storage/issues/358

`BUILD_SCM_REVISION` and `BUILD_SCM_TIMESTAMP` have no `STABLE_` prefix, so Bazel files them in `volatile-status.txt` and reports an unchanging digest for that file whatever it contains. The `expand_template` in bb-storage that turns them into the image tag is therefore never invalidated, and `_stamped.tags.txt` is generated once and reused indefinitely — on a build machine whose cache persists, every later build pushes its image to the tag produced by the *first* build on that machine.

This is where we hit it, on a self-hosted runner. In a single job the checkout was at `0250440` and `volatile-status.txt` correctly said `BUILD_SCM_REVISION 0250440`, while `_stamped.tags.txt` still held `20260407T093634Z-220cdb7` with an April mtime — four months and two version bumps stale.

The bb-storage bump is part of the fix, not incidental. That template is what expands these keys, and the pin here (`10acc76a`) predates `33530b6`, which switched it to the prefixed names. Renaming the keys alone would leave the pinned template still reading the volatile ones, so nothing would change.


matx-sjacob:
@EdSchouten please could I get a review for this one? It should be identical to https://github.com/buildbarn/bb-remote-execution/pull/244

isakstenstrom:
Looks good! I took your changes and updated the module lock file with the latest commits on main in #336.

PR #315, opened 2026-08-17
Fix crash on userinfo display

isakstenstrom:
The frontend expects the UserInfo to be a object with string values, but the backend can send nested objects as well. This change makes sure that the frontend can handle nested objects by flattening the objects and displaying them as strings first.


isakstenstrom:
Replaced by #316

PR #335, opened 2026-08-20
Fix concurrent target imports

darkrift:
This address an issue when uploading multiple invocations at once and those multiple invocations contains the same targets, all those common invocations will conflict when trying to insert the same target

### DIFF
--- a/internal/database/buildeventrecorder/BUILD.bazel
+++ b/internal/database/buildeventrecorder/BUILD.bazel
@@ -75,8 +75,10 @@ go_test(
     deps = [
         ":buildeventrecorder",
         "//ent/gen/ent/runtime",
+        "//internal/database",
         "//internal/database/dbauthservice",
         "//internal/database/embedded",
+        "//internal/database/sqlc",
         "//test/testutils",
         "@com_github_google_uuid//:uuid",
         "@com_github_prometheus_client_golang//prometheus",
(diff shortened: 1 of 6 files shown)


isakstenstrom:
Is this an actual problem you have encountered? Or is this some problem that and agent has found and fixed?

Since we have automatic retries for each batch, it should fail to insert the current batch and then succeed on the next try. Or have you found a case where this is needed?

darkrift:
I did face this multiple times when trying to upload 10 ndjson at the same time

isakstenstrom:
Okay I think I know what the problem is. The retry timeout in the BuildEventRecorder is 1 second which is fine for handling events from the BuildEventStream since the batches tend to be small, but the BEP Uploader creates a single batch with all elements it often takes longer than 1 second to save the batch.

In bb-portal we have multiple things that can fail due to other uploads modifying the database state during an upload. That is the reason we implemented a retry loop for saving batches, so that we can recover from transient errors. I discussed this with @meroton-benjamin and arrived at that we should increase the retry timeout for the BEP Uploader instead of trying to add complexity to fix every time something can go wrong when inserting into the database. A 15 second retry timeout should fix most problems when uploading ndjson files.

What do you think?

darkrift:
Fine for me if that fixes it, what if the upload fails after 15 seconds ? Should it automatically retry with a higher value ?

isakstenstrom:
> Fine for me if that fixes it, what if the upload fails after 15 seconds ? Should it automatically retry with a higher value ?

We could maybe bump it up a bit higher. I think it is fine if it has failed after a couple retries. We don't consider the bep upload a critical part of the application. It is nice for development, but we suspect that it is used very little in prod.

PR #337, opened 2026-08-24
Add BES Instance name columns to the builds and invocations pages.

tjoneslo:
In our testing we were using the portal for several projects, and using the BES Instance Name to segragate them. But there is no visibility for the instance name in the portal UI. But realizing this is not a universal requirement this hides these columns behind a configuration items in the portal configuration.

When enabled ( columnInstanceNameBuilds and/or columnInstanceNameInvocations) a "BES Instance" column is added to the corresponding page. You can seach on this value to see just those instances.

Add testing. Including adding code to run `vitest` and the vite json tests as part of the bazel build. It's not exciting testing.

I had claude / AI generate the frontend typescript code as it's not my best language.


isakstenstrom:
Nice!
I don't think it is necessary to have the columns configurable, we can just go ahead and always show them. In our internal backlog we have a ticket to make all columns configurable (but it will take quite a while to get there). In the meanwhile I think its fine to always show them.

Also, please put your test system changes in another PR. They are unrelated to the columns. That makes it easier to review.

tjoneslo:
> Also, please put your test system changes in another PR. They are unrelated to the columns. That makes it easier to review.

I extracted the vitest_test macro into #340, please review and merge before this PR.

frontend/next-env.d.ts
isakstenstrom, line 1: We don't use Nextjs anymore so this file should be deleted.

frontend/src/components/BuildsTable/Columns.tsx
```diff
@@ -143,5 +143,28 @@ export const getColumns = (): TableColumnTypeWithFilter<
     },
   );
 
+  columns.push({
```
isakstenstrom: You don't need a separate push for this. Just add it to the one above.

frontend/src/components/BazelInvocationsTable/index.test.tsx
isakstenstrom, line 1: I think these tests are a bit silly. Checking that a column exists is a trivial thing that we don't need to test.

frontend/src/components/BuildsTable/Columns.test.tsx
isakstenstrom, line 1: This test is also unnecessary.

frontend/src/components/BazelInvocationsTable/index.tsx
```diff
@@ -29,21 +32,27 @@ interface Props {
   pageInfo: PageInfo;
 }
 
+export const getTableColumns = (): TableColumnTypeWithFilter<
```
isakstenstrom: Without tests, this doesn't need to be exported.

frontend/package-lock.json
isakstenstrom, line 1: I don't see why the lockfiles have changed. This PR changes nothing that is related to the dependencies. Please remove the changes in this file, `pnpm-lock.yaml` and `npm_translate_lock_...`. Other than that it looks good.

isakstenstrom (approved):
Looks good!

PR #339, opened 2026-08-28
Fix Perfetto trace message payload

### LINKED ISSUE #338: Cannot open trace in perfetto using button
In `bb-portal` there's a button to open a trace in perfetto:

Image

However, clicking it shows an error in perfetto:
```
Oops, something went wrong. Please file a bug.

UI: https://ui.perfetto.dev/v58.3-11fbaed83

TypeError: str.replace is not a function
 - sanitizeString (../../src/frontend/post_message_handler.ts:345:2)
 - sanitizePostedTrace (../../src/frontend/post_message_handler.ts:335:2)
 - parsePostedTrace (../../src/frontend/post_message_handler.ts:136:2)
 - postMessageHandler (../../src/frontend/post_message_handler.ts:228:2)

Trace: not available (No trace loaded). Provide repro steps.
UA: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36
Referrer: <redacted>
```

Image

I do also see this error in the console:
```
post_message_handler.ts:171 Ignoring message - document not ready yet.
postMessageHandler	@	post_message_handler.ts:171
```

Image

snakethatlovesstaticlibs:
Closes #338


frontend/src/components/ProfileDropdown/index.test.ts
```diff
@@ -0,0 +1,58 @@
...
+          title: "invocation-id",
+          fileName: "profiles/profile.pftrace",
+        },
+      },
+    );
+  });
+});
+
+describe("waitForPerfettoToLoad", () => {
+  it("ignores PONG messages from another window", async () => {
```
isakstenstrom: This test has a race condition. If we wait more than 100ms (which is unlikely, but it can happen), we get the error TypeError: handle.postMessage is not a function. To make this test worth it, I think we need to mock the timers to control the execution. But I'm not convinced this is a test that is necessary.

snakethatlovesstaticlibs: Thanks for taking a look, I'm ok with that. I have removed both tests (the other one was also a bit trivial)

frontend/src/components/ProfileDropdown/index.tsx
```diff
@@ -19,7 +19,21 @@ const fetchProfileFile = async (
   return res.arrayBuffer();
 };
 
-const waitForPerfettoToLoad = async (handle: Window) => {
+export function createPerfettoTraceMessage(
```
isakstenstrom:
```suggestion
function createPerfettoTraceMessage(
```

frontend/src/components/ProfileDropdown/index.tsx
```diff
@@ -19,7 +19,21 @@ const fetchProfileFile = async (
...
+  return {
+    perfetto: {
+      buffer,
+      title: invocationID,
+      fileName: profile.filePath.path,
+    },
+  };
+}
+
+export const waitForPerfettoToLoad = async (handle: Window) => {
```
isakstenstrom:
```suggestion
const waitForPerfettoToLoad = async (handle: Window) => {
```

isakstenstrom (approved):

isakstenstrom:
Looks good, thanks!

PR #344, opened 2026-09-02
Fix cross compile build issues

isakstenstrom:
- Cross compiling freebsd, mac, windows and linux builds separetely.
- Removed unused toolchain in vite.bzl.
- Bypassed runtime toolchain usage for freebsd


tools/github_workflows/github_workflows.jsonnet
```diff
@@ -39,7 +39,12 @@ local cross_build_step = {
   local platforms = std.join(",", [
     "@com_github_buildbarn_bb_storage//tools/platforms:%s" % p for p in cross_targets
   ]),
```
mortenmj, line 41: This variable is no longer in use

mortenmj (approved):

PR #351, opened 2026-09-02
Fix behavior on missing data in CAS

isakstenstrom:
- FindMissingBlob used to identify when a blob is missing from cas
- Updated the following screens:
   - Browser (Display an error message if the data is missing) Command Action Historical Execute Response
   - Compare Action (Allows new comparison in case of missing data)
   - CAS Log Viewer (Displays that the data is missing)
   - Critical Path (Hide critical path)
   - Profile (Disable the button but display the reason through a ToolTip)


frontend/src/components/BrowserResultDescription/index.tsx
```diff
@@ -17,6 +19,38 @@ const BrowserResultDescription: React.FC<Params> = ({
...
+    [
+      {
+        hash: executeResponse?.result?.stderrDigest?.hash ?? "",
+        sizeBytes: (
+          executeResponse?.result?.stderrDigest?.sizeBytes ?? ""
+        ).toString(),
+      },
+    ],
+    browserPageParams.digestFunction,
+    executeResponse.result?.stdoutDigest !== undefined,
```
mortenmj: Shouldn't this refer to stderrDigest?

frontend/src/components/ProfileDropdown/index.tsx
```diff
@@ -73,35 +75,63 @@ function openPerfetto(profile: FileDetailsFragment, invocationID: string) {
...
+  const { exists } = useCheckDataExists(
+    profile?.digest.rev2InstanceName ?? "",
+    [
+      {
+        hash: profile?.digest.hash ?? "",
+        sizeBytes: String(profile?.digest.sizeBytes),
+      },
+    ],
+    digestFunction_ValueFromJSON(profile?.digest.digestFunction.toUpperCase()),
+    profile !== undefined,
```
mortenmj: What about the case when profile == null?

frontend/src/components/BrowserResultDescription/index.tsx
isakstenstrom: Good catch! Fixed!

frontend/src/components/ProfileDropdown/index.tsx
isakstenstrom: Fixed!

mortenmj (approved):

PR #354, opened 2026-09-02
Fix leaking of worker state between instance names

isakstenstrom:
The API allowed a user that had access to a Platform Queue to see all operations which were being executed in that queue, regardless of the instance name of the operation.


mortenmj:
Should the `killOperationsAuthorizer` get the same treatment in a follow-up?

mortenmj (approved):

isakstenstrom:
> Should the `killOperationsAuthorizer` get the same treatment in a follow-up?

What do you mean? What treatment?

PR #360, opened 2026-09-11
Bundling Loading Improvements

isakstenstrom:
- Previously, frontend bundles were loaded sequentially in a waterfall,
  impacting page load performance.
- Assets are now identified and preloaded at compile time.
- Preload headers are added to the response, allowing assets to load
  in parallel without waiting for index.html to be fetched and parsed.
- This reduces the time required to load the frontend bundles.
- Tests to verify this behavior have been included.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


pkg/frontend/embedded.go
```diff
@@ -30,20 +40,51 @@ func cacheControlMiddleware(sourceFS fs.FS, next http.Handler) http.Handler {
 	})
 }
 
+func preloadLinkMiddleware(next http.Handler, embeddedFrontendFS fs.FS) http.Handler {
+	var routeManifest []RouteManifest
+	routeMapBytes, err := fs.ReadFile(embeddedFrontendFS, "route-manifest.json")
+	if err != nil {
+		log.Fatalf("Failed to read route-manifest.json: %v", err)
```
mortenmj: Why use log.Fatalf here (and below) rather than util.StatusWrap?

PR #364, opened 2026-09-18
Add whoami-endpoint

isakstenstrom:
Add endpoint that that shows user info. Useful when debugging auth.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


mortenmj (approved):

mortenmj:
@isakstenstrom I think this shows both public and private auth metadata. Is that what we want?

isakstenstrom:
> @isakstenstrom I think this shows both public and private auth metadata. Is that what we want?

We discussed this and determined that the only case where this is a problem is when the system admin has set up the system in such a way where the private metadata contains sensitive values that not even the user themselves should see. I would think this is rare in the real world.

But I would be okay with limiting it to the public metadata if you think that would be better. What do you think?

mortenmj:
> > @isakstenstrom I think this shows both public and private auth metadata. Is that what we want?
> 
> We discussed this and determined that the only case where this is a problem is when the system admin has set up the system in such a way where the private metadata contains sensitive values that not even the user themselves should see. I would think this is rare in the real world.
> 
> But I would be okay with limiting it to the public metadata if you think that would be better. What do you think?

I think your reasoning is sound

